PR #40, opened 2022-02-22
Modified model as per previous meeting.

GraemeMuller:
- New relationships created.
- Some tables modified.
- Added some models to Django admin.


thakurpriya1990 (commented):
Can we change the 'CM_Region District' to Boranga_Region District' as CM referes to compliance Management project

thakurpriya1990 (commented):
Add the DocumentCategory as a fk in SpeciesDocument and also add species_id(SpeciesModel) as a fk in SpeciesDocument

PR #152, opened 2023-05-25
added columns, filter dependency and conservation category displayed below scientific name

amipatel6783:
-> added columns 'Family' and 'Genera' to Conservation Status referral dashboards
-> added filter dependency - district dropdown values should only populate if region selected (under Species & Communities and Conservation Status both dash)
-> conservation_category for resp. species/comm is now visible below scientific_name when 'View' that species/comm (if conservation status is 'approved' and 'applies_to_wa=True' for that species/comm)


xzzy (commented):
Sensitivity Check completed

xzzy (commented):
Sensitivity Check completed

PR #153, opened 2023-05-26
bugfix-system name not appearing on the home page

amipatel6783:
Resolved - System name was appearing as " " in the home page header and messages eg: 'Welcome to the {{system_name}}'


xzzy (commented):
Sensitivity Check Completed

PR #172, opened 2023-07-25
Ap working

amipatel6783:
test select2 clear button visibility


xzzy (commented):
Sensitivity check completed

PR #227, opened 2024-03-21
Map draw point coordinates for occurance report

ka-pr:
Tweaks to the map component to allow for drawing points additionally to drawing polygons
- some refactoring of the map component (linting, renamed variables, moved functions around, deleted comments, moved repeated code block into functions, etc.)
- restored undo/redo for the sketch coordinates of the polygon being drawn
- changed the Draw button into a toggle that pop ups a submenu which allows to chose from draw polygon and draw point
- drawn points are now saved to the OccurrenceReportGeometry model
- updated npm packages


xzzy (approved):
Sensitivity Check Completed

PR #232, opened 2024-03-26
Private Media Permissions

awf-dbca:
Added authorisation to private media documents and made some minor fixes.

All known filepaths have auth checks based on the model their documentation pertains to. for both internal and external users.

External users can access certain documents that they have submitted. Internal users may access any document as long as they belong to at least one of the required auth groups. Each document/model type has its own auth check function that use a number of existing helper functions to determine authorisation.

Some media storage was still using default storage, these instances were switched to use private.
The getPrivateFile view was adjusted to normalise and validate the file paths for security.

Note: internal document access will be subject to review and any additional document sub-directories will need to be accounted for.


xzzy (approved):
Sensitivity Check Completed

PR #233, opened 2024-03-26
Map more work on map component

ka-pr:
Added point and polygon shapefile upload


xzzy (approved):
Sensitivity Check completed

PR #235, opened 2024-03-26
Docker build and node changes

oakdbca:
- Make docker build use oim user instead of root
- Upgrade node version to 20 instead of 16 which is now deprecated (and was causing github warnings)
- Fix a front end dependency vulnerability (level: High)


xzzy (approved):
Sensitivity Check Completed

PR #241, opened 2024-03-27
Map continue working on shapefiles

ka-pr:
Added upload and backend processing of zip-archived shapefiles.


xzzy (approved):
Sensitivity Check completed

PR #245, opened 2024-03-28
Map even more work on shapefiles

ka-pr:
Changed filefield component to accept multiple input files, and added drag&drop into the map canvas for geojson files


xzzy (approved):
Sensitivity Check Completed

PR #246, opened 2024-04-02
Latest work

oakdbca:
- Add basic occurrence flora datatable
- Add default data manager to create required groups if they don't exist on system start up
- More UI tweaks
- Configure debug logs to show on console when in DEBUG mode


xzzy (approved):
Sensitivity Check Completed

PR #250, opened 2024-04-04
Map continue work on shp drag drop

ka-pr:
The map now allows for drag and drop of zipped or loose shapefiles


jmushtaq (approved):
Sensitivity check completed

PR #251, opened 2024-04-04
Latest work

oakdbca:
- Get comms and action logs working for occurrence and occurrence report
- Add form component to internal occurrence report page
- Add process action to occurrence and occurrence report when the user is an assessor and the status is with_assessor
- Enable UserActionLoggingViewset for occurrence and occurrence report so the CRUD actions are automatically logged


jmushtaq (approved):
Sensitivity check completed

PR #253, opened 2024-04-08
Latest work

oakdbca:
- Add assign to assessor functionality to internal occurrence report page
- Modify settings to only initialise sentry if not running the dev server so it doesn't log while developing locally


jmushtaq (approved):
Sensitivity check completed

jmushtaq (approved):
Sensitivity check completed

PR #264, opened 2024-04-15
First time log in check

awf-dbca:
If the user is missing contact, name, or address details they will be redirected to a first-time login page so they can fill in those details.

Works exactly the same as leases-licensing.


jmushtaq (approved):
sensitivity check completed

PR #265, opened 2024-04-15
Allow observer name across multiple reports

awf-dbca:
ObserverDetail observer_name had a unique constraint, which disallowed it from being applicable to multiple observation reports.

This has been changed so that instead the ObserverDetail model has a unique_together constraint instead.


jmushtaq (approved):
sensitivity check completed

PR #266, opened 2024-04-15
Occurrence Report Buttons/Links

awf-dbca:
Add Occurrence Report buttons work as expected - a new occurrence report is created and the user is redirected to the details page of that new occurrence report.

Links to a given occurrence report details page also now function properly.


jmushtaq (approved):
Sensitivity check completed

PR #271, opened 2024-04-17
Community OCR Lookup fix

awf-dbca:
Fixed bug where Community OCR table was being populated with ALL OCR


jmushtaq (approved):
Sensitivity check completed

PR #274, opened 2024-04-18
Add Revision User

awf-dbca:
Adds revision user to history tables. Each time a model instance is saved, the request user is recorded in the revision record.

To implement a number of serializers were modified so new instances could be saved to kwargs - essential for tracking version user.

Also adjusted some functions that would save an instance multiple times to use the no_reversion kwarg in order to prevent multiple unnecessary revision records.

Also change tz of revision modified date for history displays.


xzzy (approved):
Sensitivity Check completed

PR #276, opened 2024-04-19
Map datum more work

ka-pr:
First chunk of work on datum transformation for map geometries. Should be able to change the datum of point geometries.


xzzy (approved):
Sensitivity Check Completed

PR #280, opened 2024-04-22
Occurrence page

awf-dbca:
Tab form presented upon accessing any given Occurrence record.

Include migrations for new fields and models pertaining to occurrences.

Current work allows for details to be edited and saved. A related occurrence report table is also presented.

Land Managers field still awaiting feedback. Submit not yet implemented (if required).


xzzy (approved):
Sensitivity Check Completed

PR #285, opened 2024-04-24
Map datum transformations

ka-pr:
Adds ability to modify point coordinates from the list-of-geometries popup. Also added undo-redo for coordinate and datum editrs


xzzy (approved):
Sensitivity Check Completed

PR #286, opened 2024-04-24
Occurrence Habitat and Observation

awf-dbca:
Added the habitat and observation tabs to the occurrence view, as well as supporting endpoints, models, and serializers.

Created new models for OCC specific instances of habitats and occurrences (and related models), while also renaming OCR models. Now all relevant model names are prefixed with either OCC or OCR.

Further work will be required with regards to auto populating OCC fields based on submitted OCRs.


xzzy (approved):
Sensitivity Check Completed

PR #287, opened 2024-04-26
OCC add existing threats

awf-dbca:
Adds a modal table presenting all threats from all occurrence reports related to an originating occurrence, accessible from that occurrence's threat tab. 

Any OCR threat (that has not already been added to the OCC) can be copied to the related OCC via this table. An OCR can only be copied over to an OCC instance once. The OCC threats table now displays the original OCR that threat has come from (if any).


xzzy (approved):
Sensitivity Check Completed

PR #288, opened 2024-04-29
Threat Adjustments

awf-dbca:
Threat Numbers (of all kinds) are now displayed with a prefix to the record they pertain to e.g. OCC1 - T1

OCC threats copied from an OCR threat have the original OCR lists with the threat number e.g. OCR1 - T1

A read only OCC threat table is now available for Species and Communities, located underneath their own threat table


xzzy (approved):
Sensitivity Check Completed

PR #291, opened 2024-04-29
Map list add point

ka-pr:
Adds a button to the list of geometry to add a new point feature to the map.


xzzy (approved):
Sensitivity Check Completed

PR #294, opened 2024-04-30
OCR to OCC section copy

awf-dbca:
Each OCC sub-section for habitat and observation now has a related occurrence report table from which:

1. Users can view the values of the relevant sub-sections
2. Users can copy over the values of the relevant sub-section to that of the occurrence


xzzy (approved):
Sensitivity Check Completed

PR #295, opened 2024-04-30
History Auth Update

awf-dbca:
Meeting, occurrence documents, and threats (of all kinds) added to history auth lists - meaning members of those groups can access those history records.

(auth groups and what they can access still may be subject to changes which will need to be implemented here)

Also adjusted file auth check to return not found for files that do not exist (was returning unauthorized)


xzzy (approved):
Sensitivity Check Completed

PR #296, opened 2024-05-01
Threat Display Adjustments

awf-dbca:
Threat source for occ and ocr presented correctly (with ocr copied occ threats presenting the original ocr as the source).

Default ordering added to threat tables (by id desc).


xzzy (commented):
Sensitivity Check Completed

PR #299, opened 2024-05-01
Latest work

oakdbca:
- When adding OCR to existing OCC restrict filtering to OCCs of same species or community
- Show error messages on OCR propose approve modal
- Bug Fix: Change occurrence field from one to one to foreign key on occurrencereportapprovaldetails model
- Bug Fix: When OCR propose approve modal was closed it was resetting the form thus deselecting the default option. Fixed by removing the call that resets the form and just resetting the values on the bound object instead.


xzzy (approved):
Sensitivity Check Completed

PR #300, opened 2024-05-01
Clone OCR to OCC Functionality

oakdbca:
- Add functionality to clone the OCR to the OCC when creating a brand new OCC from the OCR.


xzzy (approved):
Sensitivity Check Completed.

PR #301, opened 2024-05-02
OCR & OCC History Columns

awf-dbca:
Adjusted OCC and OCR history columns to be more relevant to their respective records.

Removed previous name and common name values for OCC and OCR history (kept scientific/community names)

For OCC added number of reports (derived via number of related OCRs), wild status, and occurrence source.

For OCR added Occurrence (if OCR has been linked to one).

Some consideration made towards including OCR and OCC detail tables (habitat composition, observation details, etc) in reversion. Decide to hold off due to the large number of tables involved (18+ between both OCR and OCC, with revision versions compounded by OCC and OCR relations).

### DIFF
--- a/boranga/components/occurrence/models.py
+++ b/boranga/components/occurrence/models.py
@@ -3381,4 +3381,4 @@ reversion.register(OccurrenceDocument)
 reversion.register(OCCConservationThreat)
 
 # Occurrence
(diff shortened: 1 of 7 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #302, opened 2024-05-02
Latest work

oakdbca:
- Get submit of internally created OCR by internal user working
- Remove lots offiles from front end that are not used in the project (Stuff that was copied into Boranga that isn't relevant to the project i.e. approvals, compliances, vessels, vehicles, tclass, filming etc.


xzzy (approved):
Sensitivity Check Completed

PR #303, opened 2024-05-03
Misc changes

awf-dbca:
- Properly highlight selected tab in interface for object detail displays
- fixed a number of minor client-side errors
- fixed pop overs not working on page change or table expansion (pop overs should work everywhere now)
- added special pop over function to accommodate links
- adjusted history api to account for OCR - OCC relations, we want to show and OCR's OCC when generating the OCR history table, but not all the other related OCRs (this was causing an error on some OCR history tables)
- added a record display number to history details modal, indicating which instance of a model is the primary instance (OCR details will still potentially include other OCRs, so this will help identify the "primary" OCR)


xzzy (approved):
Sensitivity Check Completed

PR #304, opened 2024-05-03
Map geometry buffer

ka-pr:
Adds a map menu item to the map to run spatial operations on selected geometries. For now only able to buffer. No model has been created yet to store those processed geometries, they are added as a new feature to the map, however.


xzzy (approved):
Sensitivity Check Completed

PR #305, opened 2024-05-03
Latest work

oakdbca:
- Fix cron file so it uses oim user instead of root and uses the correct .cronenv file
- Remove no longer needed exception blocks from species and communities drf end points
- Get save and continue, save and exit working for internal submit OCR
- Add OCR External referee invite
- Get OCR referral endpoints remind, recall and resend working


xzzy (approved):
Sensitivity Check Completed

PR #306, opened 2024-05-06
Build fix

oakdbca:
- Add back in gdal apt packages to mirror kb
- Add new logic for readonly field on InternalOccurrenceReportSerializer


xzzy (commented):
Sensitivity Check Completed

PR #309, opened 2024-05-07
Map geometry convex hull

ka-pr:
Adds a couple of spatial operations to process map features (not only convex hull) and slightly improves the look and feel of the map.


jmushtaq (approved):
Sensitivity check completed

PR #314, opened 2024-05-08
OCC OCR Edit Auth

awf-dbca:
Added Client-Side and Server-Side Auth Validation for editing/processing OCRs and OCCs.

Implemented a number of rules for both, including:
 - only submitters can alter OCRs in the draft stage
 - only members of an assessor group can assign themselves to an OCR
 - only members of an assessor group assigned to an OCR can update the OCR while with assessor or with referral
 - OCCs can only be edited by members of an the OCR assessor group if they are active

In addition:
 - An OCR Assessor and Approver group have been added, distinct from their CS equivalents
 - OCCs no longer have a draft status and instead are "Active" by default
 - The Approver group for OCRs are also the the Editor Group for OCCs
 
All client-side changes work so that save/update buttons are greyed out when users. All update/workflow/assignment changes are then verified server-side.
 
OCC Location will need to be updated once complete.


jmushtaq (approved):
Sensitivity check completed

PR #315, opened 2024-05-08
Decouple and Remove Unused Django Components / Apps

oakdbca:
- Removing Approvals, Organisations, Proposals and Compliances from Boranga as they are not used
- Move any required references to components that are being kept
- Remove components folders, remove from installed apps
- Remove any references to said components in both application files and migrations
- Test full migration on fresh DB from scratch - successful


jmushtaq (approved):
Sensitivity check completed

PR #316, opened 2024-05-09
External Document Access Restrictions

awf-dbca:
Adjusted auth check for external when accessing documents.

Now for OCR documents, only those stored in /documents/ can be accessed by external users.

Access has been removed from external user to all documents for CS as none are externally submissable.


jmushtaq (approved):
Sensitivity check completed

PR #319, opened 2024-05-09
History Adjustment for Model Change

awf-dbca:
CrossReference model is being replaced by the TaxonPreviousName model.

The previous model has had its registration remove from reversion, with the new one replacing it. The Taxonomy model reversion registration has also been changed to follow the new model instead of the old.

History displays that previously display both CrossReference values and Taxonomy values in general have been adjusted. Before a species history query could return multiple taxonomies due to how CrossReference worked, now only returning one. Multiple previous names are still possible, with the latest being present in the history table.


xzzy (approved):
Sensitivity Check Completed

PR #321, opened 2024-05-10
Map layers

ka-pr:
Adds
- a new layer-switcher to the map, allowing to show/hide layers and toggle layers for editing
- a new vector-layer for processed geoemtries
- ctrl-drag-box selection for mass-selection of map features
- some refactoring of old code to get the new vector-layer working nicely (might potentially break something that has been working before)


jmushtaq (approved):
Sensitivity check completed

PR #322, opened 2024-05-10
Threat filters

awf-dbca:
Adds threat filters for OCC threats, OCR threats, S&C threats, and S&C OCC threats.

Filters include threat category, threat current impact, threat potential impact, and the observed to and from date.

OCC threats also have the threat source filter as they their threats can come from different OCRs.


xzzy (approved):
Sensitivity Check Completed

PR #323, opened 2024-05-13
Save adjustment, create fix

awf-dbca:
To accommodate key word arguments necessary for reversion and to record model instance "numbers" in each initial revision instance, the save function of a number of models were altered. 

When one of these models are saved, they are first saved with the "no_revision" keyword argument to initialise the record without saving a revision instance. The record number of the instance is then generated, and then the instance is saved again, this time with the revision creation allowed.

The problem was that is the model instance was created using the create() function, it would send through a "force_insert" keyword argument, which due to how the instance was initialised and updated would cause the second save to trigger a duplication error - the force insert would attempt to create a new record with the same PK.

While this does not affect the functionality of the models, as none of them generate records using create(), this PR still addresses this issue so that the function can be used - in case it is required for debugging or later functionality.

This PR fixes the issue by forcing any force insert keyword argument to run on the initial save for the affected models, and not the second. This has been tested - the create() function now works for all affected models and existing functionality has not changed.

### DIFF
--- a/boranga/components/conservation_status/models.py
+++ b/boranga/components/conservation_status/models.py
@@ -488,7 +488,8 @@ class ConservationStatus(RevisionedMixin):
 
     def save(self, *args, **kwargs):
         if self.conservation_status_number == "":
-            super().save(no_revision=True)
+            force_insert = kwargs.pop('force_insert', False)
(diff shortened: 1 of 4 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #325, opened 2024-05-14
NOMOS script updates

thakurpriya1990:
The management script updated according to the new API data from NOMOS. Accordingly, the backend model fields, and frontend changes done.


xzzy (commented):
senstivity check completed

PR #326, opened 2024-05-14
Map layers more layer work

ka-pr:
Some refactoring of the map component code and started implementing logic to control to which layer OCR/OCC geometry features go to.


xzzy (approved):
Sensitivity Check Completed

PR #328, opened 2024-05-15
CS URL Length Fix

awf-dbca:
Conservation Status Dashboard was failing to load due to recently added filters increasing the request url beyond allowed limit.

This PR has changed all internal requests for CS records that utilise filters to use the POST method instead of GET.

Requests that have been changed include:
 - the internal dashboards for flora, fauna, and community 
 - the exports for those dashboards
 - the referral tables within those dashboards
 - the exports for the referral tables
The filter functions have been changed to get params from POST as well.

NOT changed is the external CS dashboard - that has not breached the url limit due to using fewer filters and utilises its own filter backend.

Changes to the external CS dashboard, and all other requests, may be required as need and/or for consistency in future work.


xzzy (approved):
Sensitivity Check Completed

xzzy (commented):
Sensitivity check completed

PR #330, opened 2024-05-15
Map layers even more work

ka-pr:
Started providing props to control what the map component displays or what functionality it provides. Also some improvements to the list of map features and the styling of selected features.


xzzy (approved):
Sensitivity Check Completed

PR #331, opened 2024-05-15
System Groups Round 2

oakdbca:
- Show new contributor badge in internal CS and OCR datatables for users that have not yet had a submission either declined or approved (either CS or OCR submission), Also show this badge on the CS and OCR details pages next the contributor's name
- Add user logged in signal that adds new users to the external contributors group as long as their email is not in a blacklist
- More work getting the things editable and visible based on new user groups


xzzy (approved):
Sensitivity Check Completed

PR #332, opened 2024-05-16
Internal/External URL Conversion Changes for Email Sending

oakdbca:
- Add fuctions to convert to and from internal or external urls that can work locally (where the url does not need conversion), in dev/uat (where the internal suffix is "-internal-<rancher_cluster>" and in production (where the internal suffix is "-internal"
- Add calls to new functions to SC, CS and OCC email.py files


xzzy (approved):
Sensitivity Check Completed

PR #333, opened 2024-05-16
Species & Community Private/Public

awf-dbca:
Active Species and Community records can now be made public/private, with records sections also having public/private options.

Changes include:
- models for tracking publishing status, both of the the records as a whole and each sections
- option to make Active records public, with a radio button form presented on action to set each section as well
- display of public/private status in record tables and page, presented for Active records in the status column (i.e. Active - Private, Active-Public)
- radio button form presented for records, disabled if not public, allowing users to change which sections are public post-publication
- stricter presentation of species and communities to external users, they can only see publicly available records and can only see public sections within those records (enforced both client-side and server-side, returning non-public sections as null via API)
- External users can now view threats (if threats have been made public)
- saving a record reverts it back to private, with a message presented in the save feedback 
- history tracking added for publication status, accessible via species/community history

Design specifies that Documents are to be always private. Related Items have also been omitted for now.


xzzy (approved):
Sensitivity Check Completed

PR #337, opened 2024-05-17
S&C prevent saving if no changes made

awf-dbca:
When an S&C record is saved it makes the record private if it was public. To prevent accidental saves, a check is in place to ensure a change has been made before continuing to save.

If no change has been made, a dialogue appears informing the user that no changes have been made.

It works by maintaining a copy of the original S&C object, which itself is only updated after a successful save/refresh, and comparing the main S&C object to it before submission. The compare function has been included as a helper function.

Publishing status for an S&C record is handled by its own endpoint so has its own change check as well, and removed from the main S&C objects when they are validated (using copies so as not to alter the actual objects).

Currently exclusively applied for S&C on the client-side. May be useful to apply to other models and potentially enforce server-side as well.

### DIFF
--- a/boranga/frontend/boranga/src/components/common/species_communities/community_profile.vue
+++ b/boranga/frontend/boranga/src/components/common/species_communities/community_profile.vue
@@ -434,6 +434,10 @@ export default {
             type: Object,
             required: true
         },
+        species_community_original: {
+            type: Object,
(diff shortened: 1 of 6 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #338, opened 2024-05-20
Map layers additional tilelayers

ka-pr:
Changes the map to fetch its basemaps and optional tilelayers from URLs that may be defined in the django admin interface.


xzzy (approved):
Sensitivity Check Completed

PR #339, opened 2024-05-21
OCR workflow adjustments

awf-dbca:
A number of adjustments to OCRs per requirements.

Changes include:

 - OCRs can now be "unlocked" - once unlocked changes can be made to the OCR at which point it will revert to "with assessor" status
 - Unlocked OCRs have a "back to assessor" button presented if no changes to the OCR required but assignment to an alternative OCC is
 - When approving an OCR that had previously been assigned to an OCC, that OCC will be listed as the default choice for the "Existing OCCs" dropdown
 - Fix to OCR assignment - self-assignment was not working properly
 - Various accommodations made to support the unlock status and functionality - including changes to auth and conditional page reloads
 - Proposing an approve/decline and sending back to assessor now no longer redirects back to the main OCC/OCR dashboard
 - Observer details can no longer be deleted, but now can be removed and reinstated
 - Removing the last OCR from an Active/Locked OCC will now cause that OCC to become "discarded"
 - Moved some workflow functions (lock/unlock/close) for OCC/OCRs from the viewset to the model (auth group check at both levels)


xzzy (approved):
Sensitivity Check completed

PR #340, opened 2024-05-21
Minor Changes

awf-dbca:
Minor changes to address some test findings:

 - Submitting a species will no longer produce three history records, instead it will produce two if a change had been made (save before submit) or only one
 - History tables will now reload when their containing modal is opened
 - renamed get_occurrence_editor_group to get_occurrence_approver_group


xzzy (approved):
Sensitivity Check Completed

xzzy (commented):
Sensitivity Check Completed

PR #341, opened 2024-05-22
Viewset Mixin Change

awf-dbca:
A number of our API Viewset classes were inheriting the ModelViewSet class. The ModelViewSet includes handling for delete requests, among other things, that were not being properly handled by our own auth mechanisms beyond minimal queryset constraints and any applied permission_classes (the latter not yet implemented across the project).

This would essentially mean that (in most cases) any internal user could delete (and potentially alter) a record outside of any intended workflow and even when not part of an appropriate system group. 

This PR has adjusted the inheritance of all relevant classes to only use the super classes and/or mixins necessary for their intended functions. Most workflow, create, and update functions are handled directly in the pertaining class, so now most ViewSet classes either utilise (viewsets.GenericViewSet,mixins.RetrieveModelMixin) or (viewsets.ReadOnlyModelViewSet) - the latter used for our paginated viewsets.

Each endpoint and view has been tested to ensure no existing functionality has been affected and that everything still works as intended. The ability to delete records via these endpoints has also been reviewed - having been removed in all cases as appropriate.

Some accompanying adjustments include:
 - Minor frontend fixes to issues found during testing
 - The application of permission_classes to UserListFilterView (until now, unauth'ed users could use that endpoint)

### DIFF
--- a/boranga/components/conservation_status/api.py
+++ b/boranga/components/conservation_status/api.py
@@ -12,7 +12,7 @@ from ledger_api_client.ledger_models import EmailUserRO as EmailUser
 from openpyxl import Workbook
 from openpyxl.styles import Font
 from openpyxl.utils.dataframe import dataframe_to_rows
-from rest_framework import serializers, views, viewsets
+from rest_framework import serializers, views, viewsets, mixins
 from rest_framework.decorators import action as detail_route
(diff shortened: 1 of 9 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #343, opened 2024-05-22
Threat Adjustments

awf-dbca:
A new threat filters has been added for removed/active threats.

S&C threats have been adjusted so that if a change is made to them while the S&C record is public with threats set to public, the S&C record will revert back to private (as it does with any other change). 

Warnings have been implemented to inform the user that change to threats will revert the parent record to a private view.


xzzy (approved):
Sensitivity Check Completed

PR #346, opened 2024-05-24
Fixed OccurrenceReportGeometry object still being created with polygo…

ka-pr:
…n/point property instead of geometry


xzzy (approved):
Sensitivity Check Completed

PR #349, opened 2024-05-27
OCC Location Models

awf-dbca:
Renames Location model for OCR to OCRLocation, and adds OCCLocation models.

OCCLocation additions includes:

 - OCC geometry models and functions
 - supporting serializers
 - API endpoints
 - Basic OCC location frontend display and form
 
Currently at a state where all OCCLocation model fields can be populated via the form. The map can also be used to save geometry data though does not yet render OCC pins (related OCR pins are currently rendered). Further work will be required for full map integration.
 
 Additional work will also be required for further OCR integration.


xzzy (approved):
Sensitivity Check Completed

PR #350, opened 2024-05-27
System Groups Round 3

oakdbca:
- Ensure all calls to check system group membership pass through centralised functions
- This allows us to make sure superusers are automatically considered a member of all groups without having to change code all over the place


xzzy (approved):
Sensitivity Check Completed

PR #352, opened 2024-05-28
Bug fixes

oakdbca:
- Don't log validation exceptions to sentry
- New contributor was checking request user instead of submitter - fixed
- Fix regression that was calling self.get_success_headers on a modelviewset and when changed to a generic viewset with out revtrieve model mixin get_success_headers no longer there and causes exception.


xzzy (approved):
Sensitivity Check Completed

PR #354, opened 2024-05-28
Name search change

awf-dbca:
Main change is adding species_taxonomy to OCR and OCC models as the primary reference for the record's Scientific Name.

Species Taxonomy is directly related to the Taxonomy model, meaning all available names can be assigned to the OCC/OCR without requiring an Species record.

A relation to the Species record still exists, and for internal users the scientific name from that record (identical to that of the taxonomy-derived name) is present alongside a link to it - if one exists. 

The species record relation for OCC and OCR  are automatically created based on the provided taxonomy name. Whenever an OCR/OCC is updated or a species record is updated - the assigned taxonomy is checked - if it is in use by a species record then that species record is also assigned to the OCR/OCC (and unassigned if the taxonomy is removed from that species record).

Other minor changes included:

 - general section for species/communities no longer public (always private)
 - minor fix on OCC/OCR auth check
 - minor fix to CS list function
 - default OCC listing for OCR reassignment only presented if they share the same species/community name (with main changes of this PR accounted for)
 - previous OCC/OCR location PR included (it appears it was not previously merged)


xzzy (approved):
Sensitivity Check Completed

PR #356, opened 2024-05-29
Map proxy

ka-pr:
Added proxy integration to serve geoserver tilelayer that require authentification.


xzzy (approved):
Sensitivity Check Completed

PR #357, opened 2024-05-29
Latest work

oakdbca:
- Bug fix: Preventing assign user to OCR fixed.
- Prevent external users being added to internal system groups and vis versa
- Add manual propose delist and delist funationality to CS
- Remove management command and cron for auto delist (business doesn't require it)


xzzy (approved):
Sensitivity Check Completed

PR #359, opened 2024-05-30
Map load occurrence geometries on map init

ka-pr:
Adds occurrence geometries being loaded to the map on map initialisation.


xzzy (approved):
Sensitivity Check Completed

PR #360, opened 2024-05-31
Map geometry intersection

ka-pr:
- Unified save_ocr_geometry and save_occ_geometry into a single save_geometry fn
- save_geometry populates the intersects field in the OccurrenceGeometry model
- Added OccurrenceGeometry and OccurrenceReportGeometry to their respective admin instances


xzzy (approved):
Sensitivity Check Completed

PR #361, opened 2024-06-04
Convert Conservation Status to use taxonomy scientific name instead of species scientific name

oakdbca:
Includes typo bug fix that is prevent SC from being saved


xzzy (approved):
Sensitivity Check Completed

PR #362, opened 2024-06-04
Revert OCR/OCC name display change

awf-dbca:
Further review of requirements revealed that changes to OCR/OCC name display were not required and that Scientific Names selected for OCR/OCC records MUST have an existing species record where the respective taxonomy record is assigned.

This PR reverses most of the changes that involved maintaining a separate species taxonomy field for OCR and OCC which would automatically update the main species field - going back to the species field being manually set, selected from a list of assigned names.

Some changes from the name display change have been kept either to support the functionality for CS (which still requires allowance for an unassigned species_taxonomy field) or for better general functionality/performance (changes to cs_profile_dict endpoint, for example).


xzzy (approved):
Sensitivity Check Completed

PR #364, opened 2024-06-04
Occurrence tenure

ka-pr:
Adds an OccurrenceTenure model that is populated on (first) geometry save operation.


xzzy (approved):
sensitivity check completed

PR #373, opened 2024-06-06
External species communities access

awf-dbca:
Users do not need to be logged in to view S&C records that have been marked as public.

The dashboards had worked, but accessing specific records was still restricted.

Individual S&C records (that are marked as public) can now be accessed without authentication. This includes any visible associated conservation threats.

Other minor changes:
- processing status col header and filter removed from external S&C dash (processing status is not displayed to external users)
- removed/hidden threats are no longer visible to non-internal users
- external S&C vue file and serializer stripped of unused internal functions and values


xzzy (approved):
Sensitivity Check Completed

PR #375, opened 2024-06-06
Map query geoserver layer

ka-pr:
Restores the "map info" tool to query the geoserver for a clicked tilelayer feature and display the feature properties in a bubble. Still a bit clunky when displaying multiple additional layers, but fixing that is low priority.


xzzy (approved):
Sensitivity Check Completed

PR #376, opened 2024-06-07
Fixes

oakdbca:
- Remove unused management commands
- When approving new CS and closing previous CS set the change code (change type) for the old CS to Close
- Fix validate method on ProposedApprovalSerializer
- Remove taxonomy caching from GetScientificName as it takes longer to load the taxonomy into memory than it does to just query the database directly


xzzy (approved):
Sensitivity Check Completed

PR #377, opened 2024-06-07
Latest work

oakdbca:
- Copy new filters from CS flora to CS fauna
- Remove review_date field from CS. Now only review_due_date will be used on forms, filters and datatable columns
- Removed unused conservation plans code
- Removed search delay on SC flora datatable, seems to perform fine without that
- Remove old conservation list from CS history heading


xzzy (approved):
Sensitivity Check Completed

PR #378, opened 2024-06-07
Dev ops task 5131: Create OCR Proposals > Need to add Submitter information

oakdbca:
- Adds submitter information support to OCR
- Adds option to pre-populate observer with submitter information
- Defaults the main observer field to true if there are no main observers added yet (and false if there are any)


xzzy (approved):
Sensitivity Check Completed

PR #380, opened 2024-06-07
Section Values Update

awf-dbca:
Added a new endpoint that can list any OCR section values to view for copying in to OCC sections. Before only an id for certain lookup fields would be listed, now actual values are also listed.

Also adjusted some models with multi-choice fields - values for those fields are now established at the model level. Respective serializers for those models adjusted to leverage this.

Fixed multi-choice fields not updating when copied from OCR to OCC until refreshed - now the display will properly update.


xzzy (approved):
Sensitivity Check Completed

PR #381, opened 2024-06-10
OCC OCR copy stamps

awf-dbca:
Above the update button of each OCC section is now a display from which OCR the data for that data was sourced from, if any.

This stamp remains after update, but can be changed if data is copied from another OCR.


xzzy (approved):
Sensitivity Check Completed

PR #384, opened 2024-06-11
Associated Species NOMOS lookup

awf-dbca:
OCR and OCC Associated Species section now has a NOMOS lookup field (related_species) that can be added to and removed from, and can be copied from OCR to OCC.


xzzy (approved):
Sensitivity Check Completed

PR #390, opened 2024-06-13
Latest work

oakdbca:
- Add awaiting/completed icons to CS referred to me datatables
- Allow assessor to propose to delist approved CS, then approver can either revert back to approved or confirm the delist
- Started work on bringing CS communities detail page up to date with flora/fauna
- CS referral details page: Remove information from side bar that referees shouldn't need to see


jmushtaq (approved):
Sensitivity check completed

PR #393, opened 2024-06-13
OCR/OCC cleanup

awf-dbca:
Number of change OCR and OCC changes including:

- Removing Effective To and From Dates (not needed for either OCC or OCR)
- Adding Observation date to the OCR dashboards (with filters)
- Remove the Review Date columns from OCR (kept for OCC)
- Added Main Observer Column to OCR dashboards
- OCR Submitter and Main Observer can be searched via search bar in OCR dashboards
- Scientific/Community names have pop overs
- New OCC Name Validation - cannot submit an OCC approval proposal with an OCC name that already exists (either as an OCC or another approval proposal)

Also:

- Fixed missing serializer values preventing OCR submission
- Fixed Scientific Name lookup drop down and search


xzzy (approved):
Sensitivity Check Completed

PR #394, opened 2024-06-14
Occurrence buffer geometry

ka-pr:
First bunch of work on BufferGeometry model. Also an abstract GeometryBase class as the base for all geometry models (OccurrenceGeometry, OccurrenceReportGeomtry, and BufferGeometry)


xzzy (approved):
Sensitivity Check Completed

PR #396, opened 2024-06-17
CS Lock / Unlock and other small tasks

oakdbca:
- Remove null from many to many fields
- Bug fix in current_conservation_status property of CS model
- Bug fix in OCC allowed assessors - was returning approver member ids for assessor statuses
- Both CS species and communities status forms now save when any field in the form is changed (i.e. @change on the form element)
- Remove unused references to effective_from_date and effective_to_date (are replaced by effective_from and effective_to fields) effective_from_date and effective_to_date are still used in the proposed approve json object
- Add ability to specify that OCR documents can or can not be accessed by the submitter
- Add unlock and lock abilty for conservation status approver group when CS is finalised (declined, closed, delisted, approved)


xzzy (approved):
Sensitivity Check Complete

PR #402, opened 2024-06-17
Animal Observation Changes

awf-dbca:
- Rename Reproductive Maturity to Reproductive State
 - Change secondary sign and reproductive state from multi select fields to single select
 - Add male, female, and unknown to animal count matrix


xzzy (approved):
Sensitivity Check Completed

PR #405, opened 2024-06-18
Buffer geometry

ka-pr:
- Adds buffer geometries to the map component for occurrence
- Buffer radius is set in the frontend per occurrence geometry
- Buffer geometries are created, updated, deleted in the backend on occurrence geometry save


xzzy (approved):
Sensitivity Check Completed

xzzy (commented):
Sensitivity Check Completed

PR #411, opened 2024-06-19
Map geometry list additions

ka-pr:
- Adds a button to the geometries list to copy selected features to the current default layer
- Changes the geometries list layout to be scrollable and makes sections collapsible


xzzy (approved):
Sensitivity Check Completed

PR #412, opened 2024-06-20
Latest work

oakdbca:
- Remove region and district from community CS datatable
- Allow inviting external users to be referees for CS and OCR (Still need to add external referral datatables for the external users to be able to see the referrals)
- Made referral page more usable by puttint the referees comment field in the left column and making it sticky so they can see it when they scroll down.
- Brought OCR referral functionality up to date with CS.


xzzy (approved):
Sensitivity Check Completed

PR #416, opened 2024-06-24
Fix discard and reinstate Community record

oakdbca:
- Allow submitter to discard and reinstate communities
- Apply uniform button styling to all sweet alert 2 instances (using bootstrap classes instead of explicit colors)
- Apply new bootstrap button classes to all loading buttons or those still using old btn-default class


xzzy (approved):
Sensitivity Check Completed

PR #417, opened 2024-06-24
Discard and Reinstate Species (Flora & Fauna)

oakdbca:
- Fix discard and add reinstate action for flora and fauna records
- Replace more old button styling across the application
- Bug fix: Remove no longer used districts and regions columns from CS community datatable


xzzy (approved):
Sensitivity Check Completed

PR #418, opened 2024-06-24
Discard / Reinstate OCC

oakdbca:
- Also, add prompts to confirm create for Species flora, fauna and communities.
- Also, add prompts to confirm create for OCC flora, fauna and communities.


xzzy (approved):
Sensitivity Check Completed

PR #419, opened 2024-06-24
Map occurrence geometries copy to occ

ka-pr:
- Adds a "copy to occ layer"-link to the action column of the "related occurrence report"-table that does what the name suggests
- Slight re-modelling of the GeometryBase class to have all derived geometry classes function as generic relations for each other and themselves (for now this would be OccurrenceGeometry, OccurrenceReportGeometry, and BufferGeometry)
- Changes save_geometry fn to populate object_id and content_id of the (other) geometry a geometry is created from
- Added created_from property to GeometryBase to return the __str__-representation of the (other) geometry a geometry is created from
- Added source_of property to GeometryBase to return a list of __str__-representations of all geometries that have been created from "self", i.e. where "self" is the source of
- E.g. a geometry is created from (another) geometry when an occurrence report geometry is copied to an occurrence, or when an occurrence geometry is buffered (in the sense of creating a buffer geometry, not the spatial operation buffer)
- For some reason when a model functions as a generic relation to itself, a GenericRelation field seems to always return an empty queryset. This can easily be worked around with properties that return what the generic relation (created_from) and the inverse generic relation (source_of) would return


xzzy (approved):
Sensitivity Check Completed

PR #421, opened 2024-06-25
Occurrence report table copied to occ column

ka-pr:
Adds a "Copied to OCC" column to the OCR table.


xzzy (approved):
Sensitivity Check Completed

PR #422, opened 2024-06-26
Latest work

oakdbca:
- Bug fix: Unlock/lock CS typo preventing fields being edited when unlocked
- Copy across external referee invite code from CS to OCR for the external referee invite functionality on the OCR workflow
- Bug fix: Contact field was disappearing when contact property was empty on view observer modal.
- Bug fix: Remove old region and district filtering code form species and communities filter backends.
- Bug fix: Unable to unassign the assigned approver when status was ready for agenda
- Agenda view link now opens in new window (link has icon to indicate this will happen)
- Bug fix: Copy pasta issue with immature and ripe fruit fields. Fixed.
- Bug fix: Allow external and internal contributors to see observer contact details.
- Add a confirmation when adding new occurrence reports (external user)
- Add mandatory field indicators on most forms (still finalising OCC)
- Add change type column to Aganda datatable
- Fix filtering bug when adding CS to meeting with status ready for agenda
- Modify is_internal to return true for users that pass the dbca domain check and are referees
- Lock down meetings permissions to CS assessors, CS approvers and Read Only Users.


xzzy (approved):
Sensitivity Check Completed

PR #424, opened 2024-06-27
Occurrence tenure datatable filter

ka-pr:
- Adds` filter for feature id, status, vesting, and purpose to the occurrence tenure datatable
- Vesting filter is only a dummy until we figured out what vesting is supposed to be


xzzy (approved):
Sensitivity Check Completed

PR #425, opened 2024-06-27
Observer and Contacts constraint adjustments

awf-dbca:
Adjusts models for OCR observer details and OCC contact details to no longer enforce a unique constraint and instead utilise validation. Previously the use of constraints was adequate for the use case - OCRs/OCCs should not have multiple observers/contacts with the same name. 

However with the addition of the combine functionality, some adjustments were needed to account for scenarios where discarded contacts could be disregarded for purposes of copying contact details from a selected combined OCC. 

This change allows for one visible contact/observer name per OCC/OCR while allowing as many non-visible (discarded) records.


xzzy (approved):
Sensitivity Check Completed

xzzy (commented):
Sensitivity Check Completed

PR #426, opened 2024-06-27
Permissions Work

oakdbca:
- Enable default permission class for entire api and set to require user to be authenticated
- Add 'AllowAny' permission class to those few apis that are fully pubically accessible
- Create new permission classes for all meetings apis
- Create new permission classes to cover all conservation status apis
- Remove two unused api end points
- Bug fix: SubmitterInformationSerializer was not serializing contact information for the submitter of the CS/OCR. Fixed.
- Modify conditions for who sees CS, OCC and OCR datatables on the front end based on group membership


xzzy (approved):
Sensitivity Check Completed

PR #428, opened 2024-06-28
CS get_queryset fixes

oakdbca:
- Make all new permission classes return False if user is unauthenticated
- Make all new permission classes return True if user is superuser
- Tweak CS get_queryset methods to return correct records based on group membership
- Only allow CS approvers to create meeting records
- Add prop to form section so that sub titles are muted by default. Can be overriden as desired.


jmushtaq (approved):
Sensitivity check completed

PR #429, opened 2024-06-28
Latest work

oakdbca:
- Change Lodged on label to Submitted on in submission component
- Change OCR filtering to filter on lodgement_date instead of reported_date (reported date is model creation timestamp)
- Change reported_date column to lodgement_date on all OCR datatables
- Further modification of mandatory field indicators
- Add cache to group membership function as there are a lot more permission checks being done now and was leading to slow performance
- More granual checks for who seens which internal primary menu items
- Add missing number_of_observers and has_main_observer fields back into OCR serializers (accidentally removed during merge at some point)
- Tweak observer modal to wait until it has the record loaded (view/update) before showing the modal


xzzy (approved):
Sensitivity Check Completed

PR #430, opened 2024-06-28
Bug Fixes

oakdbca:
- Check superuser_ids is not None before iterating
- Add code back in that was accidentally removed from belongs_to function
- Cache result of is_conservation_status_referee and is_occurrence_report_referee


xzzy (approved):
Sensitivity Check Completed

PR #436, opened 2024-07-01
Fixed an exceptiopn in ListInternalOccurrenceReportSerializer that oc…

ka-pr:
…curred when an ocr has been copied from another ocr


xzzy (approved):
Sensitivity Check Completed

PR #440, opened 2024-07-02
Occurrence report geometry permissions

ka-pr:
Adds permissions for assessor for ocr geometries with status With Assessor/Referral, and adds a site layer to the map that is empty for now.


xzzy (approved):
Sensitivity Check Completed

PR #441, opened 2024-07-02
Combine OCC Sites + Site Geometry

awf-dbca:
Adds combine functionality for the sites table, where rows from selected occurrences can be chosen to be copied in to the main occurrence instance. Includes restrictions based on the site name.

Also added site geometry values which can be edited via form (implemented) and map (not yet implemented).


xzzy (approved):
Sensitivity Check Completed

PR #443, opened 2024-07-03
latest work

oakdbca:
- Change name of CS workflow from 'Intermediate' to 'Immediate'
- Add DeleteProtectedModelAdmin class that can be extended from to only allow superusers to delete items in django admin
- Add and improve actions available for SC image (upload, replace, reinstate and discard) - will wait until tested to copy across to work for communities backend.


xzzy (approved):
Sensitivity Check Completed

PR #446, opened 2024-07-04
OCC Combine Tenures

awf-dbca:
Tenures are now integrated in to the OCC Combine functionality. Tenure records are moved, not copied, similar to threats. Unlike other tables, tenures have a non-direct and conditional relation with their respective OCC - so are handled differently to other tables and depending on their status.

Tenures can be combined if they are historical, or if they are current and a current tenure with the same feature id is not in recipient occurrence.

Some other changes have also been made for the combine form in general, more combine table fields are searchable and popovers have been fixed in various places. Site datum names are also derived from a common search function based on the srid as a property.


xzzy (approved):
Sensitivity Check Completed

PR #450, opened 2024-07-05
Added tenure_area_point_on_surface property to help highlight cadastr…

ka-pr:
…e areas that are bent in such a way that centroid points would be placed outside the areas shape


xzzy (approved):
Sensitivity Check Completed

PR #451, opened 2024-07-05
Minor Adjustments and Site Geometry Interface

awf-dbca:
- Renamed Coordination Source to Coordinate Source for OCR/OCC
- Added kingdom column to related species table for OCR/OCC
- Discarded sites no longer show on map
- Sites coordinates and Datum can be change via map
- Datum is stored in-line with map requirements when submitted via site table form


xzzy (approved):
Sensitivity Check Completed

PR #454, opened 2024-07-05
Latest work

oakdbca:
- No longer return true from is_internal for internal referees that are not in any internal boranga system groups
- Simplify access to internal menu items now as a result of above change
- Enable django settings to pass protocol scheme through to django (i.e. https)
- Add number of occurrences, extent of occurrences and actual area of occupancy calculated fields to species form
- Make Area of Occupancy (2km x 2km) a manual field


xzzy (approved):
Sensitivity check completed

PR #455, opened 2024-07-08
New GIS Calculations for Species Communities

oakdbca:
- Add new calculations for flora group type that use the buffers of the points (since currently fauna does not have polygons)
- Copy functions / serializer values and html/js from species to communities.


xzzy (commented):
Sensitivity Check Completed

PR #457, opened 2024-07-08
Add Count date to OCR Plant count and Animal observation and make it available to copy it into OCC.

preranaandure:


xzzy (approved):
sensitivity check completed

PR #458, opened 2024-07-08
Map identify geometries

ka-pr:
Slight update to how geometry mouseover fearture toasts are created
- Added a mapping-dictionary to the layer definitions to control which properties are to be displayed under which name
- Added a function to compile those properties with re-usablity in mind for future geometry-export tasks
- Changed the toast to preferrably appear in the opposite corner of the map relative to the mouse cursor, to not obstruct the cursor
- Added a close button to the toast window

### DIFF
--- a/boranga/frontend/boranga/src/components/common/component_map.vue
+++ b/boranga/frontend/boranga/src/components/common/component_map.vue
@@ -1136,43 +1136,74 @@
                     </div>
                 </div>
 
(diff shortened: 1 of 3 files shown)


xzzy (approved):
Sensitivity Check Completed

xzzy (commented):
Sensitivity Check Completed

PR #460, opened 2024-07-09
Latest work

oakdbca:
- Bug fix: Permission denied when submitter attempts to upload documents to CS
- Bug fix: error with null epsg_code value when saving Occurrence geometries (removed epsg_code from location serializers as no longer required)
- SC Distribution Calculated Field Updates:
  - Filter only geometries associated with occurrences that are active or locked
  - Add code to support Point fields in convex hull calculation
  - Changes to Meeting Room section on meeting form to support new archiving functionality (to be tested on one model by BA/PM prior to rolling out to other models)


xzzy (approved):
Sensitivity Check Completed

PR #461, opened 2024-07-10
Latest work

oakdbca:
- Bug fix: allow assessor propose delist without having to assign to themselves
- Add auto calcuations of animal count to front end
- Replace total_count fields on animal observation models with property that does the calculation


xzzy (approved):
Sensitivity Check Completed

PR #462, opened 2024-07-10
Occurrence Permissions

awf-dbca:
Moved existing auth checks, including those that account for processing status and auth groups, in to permission classes.

New Permission Classes include:

- OccurrenceReportPermission
- ExternalOccurrenceReportPermission
- OccurrenceReportObjectPermission
- ExternalOccurrenceReportObjectPermission
- OccurrencePermission
- OccurrenceObjectPermission

with some adjustments to IsOccurrenceReportReferee.

No change to intended functionality has been implemented, but all auth functions are now consolidated. The exception being read allowance, which now requires group membership instead of just being an internal user.

"Object" permissions account for foreign relations which derive their CRUD allowances from their "parent" OCC/OCR.

Also changed certain get/post requests in to patch requests as appropriate for their function.


xzzy (approved):
Sensitivity Check Completed

PR #464, opened 2024-07-10
Minor Fixes

awf-dbca:
- Switched and included labels for lat, long in site table and form
- Fixed OCR/OCC assoc. species table being overwritten by save/update to whatever it was on load
- Added accommodation to process shape files EP for internal OCR permission classes


xzzy (approved):
Sensitivity Check Completed

PR #465, opened 2024-07-10
Bug fixes

oakdbca:
- Show archived as Yes or No in text rather than green tick in django admin
- Add check that assessor_mode exists on cs objects before checking value
- Update queryset filtering for CS documents to be more accurate


xzzy (approved):
Sensitivity Check Completed

PR #466, opened 2024-07-10
Map geometry list show hide geometries

ka-pr:
- Adds show/hide of individual occurrence report geometry features from the map's list of geometries
- Adds Show on Map and Hide On Map action items to the related-occurrence-report Action column
- Removes the close button from the mouseover-feature toast, because it is too elusive
- Added back area information to feature toast


xzzy (approved):
Sensitivity Check Completed

PR #468, opened 2024-07-11
Bug fixes

oakdbca:
- Bug fix: Link for historical CS when splitting was going to list view now goes directly to details page
- Add centralised styles for nav pills so that they all look the same (e.g. OCC combine modal)
- Modify authorisation checks for CS and OCR documents to be more thorough and accurate
- Exclude recalled referrals from the list returned from referred_to_me actions
- Add to action log when making species historical
- Bug fix: copy pasta issue in __str__ method of OccurrenceTenureVesting model
- Bug fix: Add missing field approval_level to SaveCommunityConservationStatusSerializer (explains why it wasn't auto saving correctly
- Resolve a bunch of TODO comments


xzzy (approved):
Sensitivity Check Completed

PR #469, opened 2024-07-12
History Expansion

awf-dbca:
The following models now have full reversion history tracking and frontend interfaces:

- OCRObserverDetail
- OCCContactDetail
- OccurrenceSite
- OccurrenceTenure

Related models to both Occurrence and Occurrence Reports are now also tracked in their respective main model history tracking.

This means every OCC/OCR field will be preserved and presented, such as those belonging to OCR/OCCLocation, HabitatCondition, AssociatedSpecies, etc. Some details from these fields are limited as reference fields are NOT retrieved on related models at this time (only the main model).

To avoid saving too many history records, the tracking of OCR records with every OCC has been removed. OCR-OCC historical relations can still be tracked via the OCR history.


xzzy (approved):
Sensitivity Check Completed

PR #470, opened 2024-07-12
Occurrence Adjustments

awf-dbca:
Occurrences can now be re-opened by Occurrence Approver group members.

Occurrence Reports change their status to with Assessor when they no longer have any active referrals after a referral recall.


xzzy (approved):
Sensitivity Check completed

PR #471, opened 2024-07-15
Latest work

oakdbca:
- Remove mandatory field indicators when SC form is readonly (e.g. when being viewed by public)
- Query optimisation for some SC viewsets
- Fix tab loading issue on OCR when tabs are clicked too quickly
- Resove some silent js errors on OCR / OCC when refs were being referred to before they were mounted
- Make sure submitter information form is always disabled for referees
- Remove documents datatable from main OCR profile tab for referees as they have access to the documents tab (unlike external contributors)
- Replace is_staff property on UserSerializer with is_internal as is staff was not being used and is_internal was needed
- Add auto save to OCR profile form
- Don't recreate the map component in save_before_submit as was causing map loading error (will monitor this change for regressions)


xzzy (approved):
Sensitivity Check Completed

xzzy (commented):
Sensitivity Check Completed

PR #472, opened 2024-07-16
Map layer ordering and more

ka-pr:
- Map layers now are ordered: buffer < ocr < occ < site
- Added django-colorfield: https://pypi.org/project/django-colorfield/
- Added color and stroke fields to geometries
- Added colorpicker to geometry admin panels
- Added opacity fields to buffer and occ geometries
- Layer opacity values for buffer and occ are now saved


xzzy (approved):
Sensitivity Check Completed

PR #473, opened 2024-07-17
Latest work

oakdbca:
- Bug fix: name: property on wild status column was not referencing a field in wild status but wild status directly which is a foreign key and thus the search error. Now referencing wild status name.
- Bug fix: Remove code that was only showing department file numbers based on distribution public boolean
- CS workflow change: Only show request amendment and Propose Ready For Agenda actions when status is with assessor. If status is with referral then assessor must either wait for all referrals to be completed or recall any outstanding referrals before they can request amendments or propose ready for agenda.
- SC Workflow update: Now when a species is made historical due to the split action the associated approved CS is not automatically closed (this will be done manually by DBCA business users
- Bug fixes when adding/editing observers
- Bug fix: Now permit types are filtered based on group type (flora, fauna or community) on OCR and OCC forms.
- Update CS amendment requests so they work the same as OCR amendment request (i.e. show a list of clickable files)
- Add new 'View Document' modal and action for CS and OCR so that users without edit permissions can still view all the fields especially in the case the description field contains a long url that they would need to be able to select
- Remove Species django admin (was only being used for dev purposes - users should be able to do everything they need via the front end
- Remove OCR and OCC django admin (used for dev purposes only users - users should be able to do everything they need via the front end)


xzzy (approved):
Sensitivity Check Completed

PR #475, opened 2024-07-18
Latest work

oakdbca:
- Replace font awesome spinners with bootstrap spinners for consistency across the application
- Save OCR before showing Propose Approve modal
- Modify permissions for who can see Meetings
- Changes to CS filtering 1. When filtering 'all' don't show discarded CS. 2. Add option to specifically show discarded CS 3. If the user is an internal contributor add a filtering option for 'Discarded by Me'
- Implement SystemEmailGroup to store lists of notification emails that the business users can add/edit/remove as required to be used to add the to the email CC list for certain actions
- Create data fixture to populate SystemEmailGroups


xzzy (approved):
Sensitivity Check Completed

PR #477, opened 2024-07-18
Map export boundary and attributes

ka-pr:
- Buffers can now created from multipolygon occ geometries (task 5150)
- Map now allows to export from selected collection, rather than from the default layer
- Geometries now have updated-date and drawn-by fields (except for buffer geometries, since they technically are now drawn by anyone)
- When a new tenure entry is created, existing manual data from the most recent historical tenure area is re-used if possible (task 5508)


xzzy (approved):
Sensitivity Check Completed

PR #480, opened 2024-07-19
Occurrence tenure purpose vesting label and code

ka-pr:
- Occurrence tenure purpose & vesting now have a label field instead of name, but still missing code
- TileLayer class now has a permission class
- Geometries (except for buffer) now track the id of the user who last updated the geometry
- Fixed opacities being undefined when there are no geometries


xzzy (approved):
Sensitivity Check Completed

PR #483, opened 2024-07-22
Request User Action Logging

oakdbca:
- Whenever a user action is logged against an object also create an action log for the request user that performed the action
- Resolve some more TODO comments
- Make validation for CS internal and external contributors the same
- Add client side validation for OCR for internal and external contributors


xzzy (approved):
Sensitivity Check Completed

PR #484, opened 2024-07-23
Occurrence tenure code field

ka-pr:
- Occurrence tenure purpose and vesting now have a code field
- Tenure and history tenure table now show purpose and vesting code instead of label
- Fixed issue with unsetting feature model after saving
- Resolved todos in map component


xzzy (approved):
Sensitivity Check Completed

PR #486, opened 2024-07-23
Fix/save set feature id on cloned feature

ka-pr:
Fixes save issue that resulted from setting a wrong feature id on a cloned feature


xzzy (approved):
Sensitivity Check Completed

PR #487, opened 2024-07-24
Map geoserver and multigeometries

ka-pr:
Fixes sql statements to treat multipoint geometries in the same way as points


xzzy (approved):
Sensitivity Check Completed

PR #489, opened 2024-07-24
File Extension Whitelist and Check

awf-dbca:
New model for file extension whitelist, with fixtures (potentially wip) included.

Additional extensions can be added via whitelist. Compression/Archive formats can also be added by checking the compressed field, though only zip, 7z, and tar archive formats are supported (but can still be used with other compression formats where applicable).

Every uploaded file is checked, in memory, before being saved to disk. Archive format contents are also checked. If the provided extension(s) is not in the whitelist or an archive contains another compressed/archived file then the upload is rejected.


xzzy (approved):
Sensitivity Check Completed

PR #490, opened 2024-07-25
Vue errors + subprocess removal

awf-dbca:
Fixed a number vue errors from missing props via the S&C split, combine, and rename functions.

Removed subprocess calls to copy documents to new dir after split, combine, and rename functions.

Subprocess calls were removed as they did not work reliably (some instances where directories would not be created to certain files would not be copied). 

As of this change copied S&C document records now point to the original file in its original upload location, under the sub-dir id of the original record. This is the same for OCC documents.

IF S&C or OCC documents are ever to become available to users on a record-by-record basis, then the files will either require copying/moving or the parent id must be otherwise obtainable to auth the user's access. As of now, neither use case appears to be required. If copying/moving is to be done, should be ideally conducted via python over subprocesses if possible.

### DIFF
--- a/boranga/components/species_and_communities/api.py
+++ b/boranga/components/species_and_communities/api.py
@@ -2,7 +2,6 @@ import json
 import logging
 import mimetypes
 import os
-import subprocess
 from datetime import datetime
 from io import BytesIO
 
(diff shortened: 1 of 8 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #493, opened 2024-07-26
Latest work

oakdbca:
- Make sure text emails inherit from text templates
- In comms log datatable prevent links from scrolling to the top and prevent popover from expanding beyond screen width
- Pin gdal to version that works with new build
- Update Boranga build to latest standards (Ubuntu 24.04, PythonScheduler, Azcopy)


xzzy (approved):
Sensitivity Check Completed

PR #494, opened 2024-07-29
Submitter Category Filtering based on User Type

oakdbca:
- Submitter categories can be assign as being for internal users or external users
- Internal users and those in the internal contributors group see the internal users list of categories
- Everyone else sees the external users list of categories


xzzy (approved):
Sensitivity Check Completed

PR #498, opened 2024-07-30
Security Review - unauthenticated, external, read-only (server-side)

awf-dbca:
Adjustments and fixes to permissions, viewsets, etc. So far tested with unauthenticated users, external users, and internal read-only users. 

Fixes involve ensuring endpoints are not accessible unintended recipients, and accessible to intended recipients. Some adjustments to previously established groups made to accommodate intended workflow (i.e. internal contributor has been adjusted to be mostly the same as external contributor and is no longer actually "internal"). Sensitive data (mainly user account names and emails belonging to people other than the request user) have been removed for external users.

Some permission adjustments made so that internal users can contribute, other adjustments made so that referees are unable to loophole access to unrelated OCRs/CS records.

Tests included document access and history access (with adjustments made as required).

Changes unrelated to auth include some minor bug fixes found during testing and adding default pagination.

STILL TO BE DONE:

 - client side adjustments for reviewed auth groups - read only and external users can access links/buttons that result in an error due to being unauthorised (e.g. Assign to me for read only, history links for external).
 - internal workflow restriction tests, testing intended workflows for all internal user groups
 - superuser access, ensuring the superuser can at least do everything that all other groups can


xzzy (approved):
Sensitivity Check Completed

PR #499, opened 2024-07-30
Addressing TODOs

awf-dbca:
TODOs addressed - 5 dismissed, 7 changes made, 7 replaced with notes, 11 TODOs remaining

TODO: Would be nice if the document id was included in the upload path to simplify this query
Not necessary, would require require a lot of change for little return

TODO: This 'document id' is actually the conservation status id. Consider renaming these variables
variable, and others like it, renamed

TODO: determine what this actually changes
referred to validate_map_files - it validates and then applies uploaded shape files

TODO: Unused variable here. Use or remove.
group type not used, was removed

TODO group auth
handled both by permissions and is_internal check (which now operates exclusively for internal groups)

TODO add version_user when history implemented
version user history added

TODO add version_user when history implemented
version user history added

TODO: current requirement task allows assessors to unlock, is this too permissive?
was confirmed to be allowed

TODO what is this for? is it needed?
removed in favour of NOTE: this and OCCLocation have a number of unused fields that should be removed (ln 1712)

TODO: possibly remove buffer radius from location models when we go with the radius being a property of the geometry
covered by previously mentioned note

TODO: Don't have these url names yet
appears to have never been addressed, may not be required (possible usage by map component?), replaced with NOTE: review if needed

TODO enable when we have history
enabled (force insert check for tenure records)

TODO: review - may not need this (?)
removed pertaining code (property for record reversion ids, did not work nor is required)

TODO: Update to use the correct endpoint
appears to be the correct endpoint

TODO Completely redo the permissions for actions on this page
changed to note, might be addressed during security review if able

TODO add validation for fields required before submit
changed to note, might be addressed during security review if able (and if even needed)

set to 200 due to the tab fade (TODO: consider better handling of this)
changed to note for now

TODO Ash - move this to dedicated save and replace with setting to private
species_split_save is (and can only be) called on an existing species record (split/combine species are preemptively created) - so copying the publishing status in unnecessary as it is already done on creation (defaulting to private)

TODO edge case, consider not using POST for process_shapefile_document if possible
replaced with NOTE replace/remove when process_shapefile_document is reworked

### DIFF
--- a/boranga/components/occurrence/api.py
+++ b/boranga/components/occurrence/api.py
@@ -2155,7 +2155,6 @@ class OccurrenceReportViewSet(
     def validate_map_files(self, request, *args, **kwargs):
         instance = self.get_object()
         validate_map_files(request, instance, "occurrence_report")
-        # TODO: determine what this actually changes
         if instance.processing_status == OccurrenceReport.PROCESSING_STATUS_UNLOCKED:
             self.unlocked_back_to_assessor()
             instance.save(no_revision=True)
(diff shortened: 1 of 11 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #501, opened 2024-08-01
Archivable Models - Priority 3

oakdbca:
- Make many models archivable
- Add archiving functionality to their django admins
- Add code to front end to allow users to change the value of a django lookup that has been archived
- Rework committee members selection mechanism on meeting details page to make it work smoothly with archiving of committee members


xzzy (approved):
Sensitivity Check Completed

PR #502, opened 2024-08-01
Workflow and Display adjustments

awf-dbca:
- Adjusted save buttons, history table, and threat table to be available when S&C record active/draft and not when discard/historical (with reinstate/reopen options available)
- Added server-side status checks for uploading documents, threats, and images to S&C records (discarded/historical must be resinstated/reopened first)
- External permissions for OCR and CS include internal groups now (Read and Create already available, but not update) 
- Fixed readonly for S&C so that it will return true for unauthorised users

Some minor client-side issues exist for CS, deferred for later maintenance.

Creating new S&C records via split/rename/combine remains unregulated via processing status on the server-side - should be addressed once aforementioned processes are reworked.

### DIFF
--- a/boranga/components/conservation_status/permissions.py
+++ b/boranga/components/conservation_status/permissions.py
@@ -97,10 +97,15 @@ class ExternalConservationStatusPermission(BasePermission):
         if request.user.is_superuser:
             return True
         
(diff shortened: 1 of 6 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #503, opened 2024-08-02
Latest changes

oakdbca:
- Log appropriate email communications against the user as well as the object (CS / OCR)
- Add confirmation prompts for creating CS (internal and external) as was previously missed
- Remove more unused files
- Replace remaining font awesome spinners with bootstrap spinners
- Update dtPopoverSplit to work the same as dtPopover. I.e. 'hover for more', smaller link and clicking it does not scroll to the top of the page.
- Fix minor CS front end issue where assign to me link was shown to in appropriate users
- Hide currently unused 'Conservation Attributes' from CS Publishing
- Add support for extra large modals (And make use of them for SC split, combine and rename. And Occurrence combine modals)
- Fix bug with is referee check returning incorrect value due to overgeneralised cache key name
- Fix bug with SC rename not sending CC emails to appropriate users
- Type fix for plant count label
- UI overhaul of SC Split, Combine and Rename modals (including adding some missing validation and confirmation prompts)
- For SC Split make sure the lookup for the new species being split into don't already have a boranga species record
- For SC Combine make sure the lookup for the species being combined into doesn't already have a boranga species record
- For SC Combine make sure the lookup for other species to be combine into new species only returns those records that do have an existing boranga species record


xzzy (approved):
Sensitivity Check Completed

PR #505, opened 2024-08-06
Map plausibility polygon

ka-pr:
- Adds a model of "plausibility polygons" intended to heuristically check in the frontend end how many tenure area potentially intersect with a given occurrence geometry
- The calculation is based on average area of a tenure within a plausibility polygon, and intersected tenures count as warning or error values
- The map component issues a warning or rejects occurrence geometries that are projected to intersect with too many tenures
- The check is calculated on finished-drawing, transformation-end, and modify-end events
- Also adds a resultType "hits" geoserver query to the backend to raise an exception when too many tenure areas intersect
- Plausbility geometries are cached
- Adds a json fixture
- Raises DATA_UPLOAD_MAX_MEMORY_SIZE setting from 2.5 to 20 MB


xzzy (approved):
Sensitivity Check Completed

PR #511, opened 2024-08-08
Wmts

ka-pr:
- Adds Web Map Tile Server (WMTS) support
- ol TileLayer can now be created from WMS and WMTS sources
- Source parameter options may be fetched from a GetCapabilities call


xzzy (approved):
Sensitivity Check Completed

PR #512, opened 2024-08-09
Fixed inner join to left join

ka-pr:
Changed an inner join to left join for the three occurrence report sql statements


xzzy (approved):
Sensitivity Check Completed

PR #513, opened 2024-08-09
Archivable Models, Blacklist Fixes and Migrated From ID

oakdbca:
- Make more models archivable (only a few edge cases remaining from the original list of models to support archiving)
- Bug fix: Fixed some issues with blacklisting users from external contributors group
- Add migrated from ID field to OCC and OCR forms and datatables


xzzy (approved):
Sensitivity Check Completed

PR #514, opened 2024-08-09
Map transform coordinates from map to user input

ka-pr:
- Moving point features on the map now correctly updates the user input coordinates in the geometry list if they are in a different crs than wgs-84
- Fixed api endpoint for available crs srid not returning all crs of all geometries, which resulted in some geometries being initialised in the geometry list with an empty crs select value if the crs srid was different from 4326


xzzy (approved):
Sensitivity Check Completed

PR #517, opened 2024-08-14
Latest work

oakdbca:
- Add copy OCR feature (pending business feedback)
- Add action logs for rename, split and combine species (appropriate logs show in both parent and child action log)
- Many small bug fixes


xzzy (approved):
Sensitivity Check Completed

PR #521, opened 2024-08-29
OCR Bulk Import Concept Fixes

oakdbca:
- Mainly to allow saving of a schema with no columns (which is required for creation)


xzzy (approved):
Sensitivity Check Completed

PR #523, opened 2024-08-30
OCR Bulk Import Schema Changes

oakdbca:
- Make sure every import includes the migrated from id column by default
- Fix the version number assignment to work for the first schema for each group type
- Add unique together contraints to prevent the same field and/or column name being included in a schema twice
- Other small tweaks


xzzy (approved):
Sensitivity Check Completed

PR #526, opened 2024-09-06
Latest Changes

oakdbca:
**OCR Cleanup**
  - Remove unused OCR fields / models

**OCR Bulk Import Tool**
  - Add ability to exclude certain fields from the list on a per model basis
  - Filter out models from the content type dropdown if all their fields are already included in the schema
  - Make Selected Column Details panel sticky top so it's still visible when scrolling down when there are lots of columns
  - Add full traceback to log output when bulk import encounters unexpected exception
  - Allow adding all fields for whole model or just mandatory fields for whole model at once rather than having to select single columns
  - Importer is now successfully importing multiple models per row (Tested OCR with OCRVegetationStructure)
  - Importer is now successfully associating a species with the OCR based on a lookup that spans multiple models (i.e. taxonomy__scientific_name


xzzy (approved):
Sensitivity Check Completed

PR #529, opened 2024-10-17
Ocr bulk import

oakdbca:
- Add default value of false for is_emailuser_column when adding new columns to schema
- Allow validate schema on front end to handle multiple exception formats from backend
- Add loading spinner while large schemas are loading
- Add code to the validate schema backend so that only species/commuties that have one or more occurrence are used as sample data when validating a schema (Otherwise validation will fail when trying to get a sample value for the linked occurrence)


xzzy (approved):
Sensitivity Check Completed

PR #530, opened 2024-10-17
Ocr bulk import

oakdbca:
- Filter out some unwanted models from showing in the list of models that can be used when generating a schema
- Now when adding new columns to a schema they will automaticaly be prefixed with an abbrevation of the model that they belong to


xzzy (approved):
Sensitivity Check Completed

PR #532, opened 2024-11-05
Dev Ops task 7176: OCR - Referred to me Bug

oakdbca:
- Datatable error was caused as the permission check returns false if the user has no occurrence reports referred to them. To fix we don't show the ocr referred to me datatable unless the user actually has one or more occurrence reports referred to them.


xzzy (approved):
Sensitivity Check Completed

PR #536, opened 2024-11-20
Documents business rules

oakdbca:
Logic pertaining to documents changed across the entire application to support the following business rules:

1) Document database records are never deleted, only ever archived (Records compliance and auditability)
2) Files uploaded by external users (external contributors) can be removed from our filesystem if the parent record (CS or OCR) has not yet been submitted (Privacy Compliance)
3) Once the parent record has been submitted, when the external user discards a document record, files are not removed from our filesystem but are kept as official records (Records compliance and auditability)
4) Discarded records are always visible to boranga internal staff with adequate permissions (Records compliance and auditability)


xzzy (approved):
Sensitivity Check Completed

PR #539, opened 2024-11-26
OCR Bulk Import Bug Fixes and Default Schemas

oakdbca:
- Bug fix: Make sure the column has appropriate abbreviations added when adding more than one column at a time.
- Fix missing allow blank option when defining schema options for TextField, ForeignKey, M2M etc.
- Truncate long column names so layout is not broken.
- Add fixture for default ocr bulk import schemas.
- Add fixutre for default ocr bulk import schema columns.


xzzy (approved):
Sensitivity Check Completed

PR #540, opened 2024-11-26
Resolve issue with content type ids in fixture not matching other environments

oakdbca:
-Use natural foreign keys instead of plain ids as content types don't match other environments.


xzzy (approved):
Sensitivity Check Completed

PR #546, opened 2024-12-30
Move collect static after vue build

oakdbca:


xzzy (approved):
Sensitivity Check Completed

PR #549, opened 2025-01-06
Boranga UAT Tasks

oakdbca:
Task | 7655 | CS Field edit - make Other Conservation Assessment a Django List
Task | 7656 | CS Form edit - move Conservation Criteria above WA Priority List
Task | 7657 | CS field edit - double the character limit on Comments field
Task | 7677 | Bug SC - S&C form should not be editable from the View action
Task | 7678 | UX SC - add an Edit action when in View only mode
Task | 7680 | UX SC - keep form open after Activating
Task | 7682 | SC Data validation for Last Curation date
Task | 7683 | SC Last Curation date error message
Task | 7684 | Bug SC Districts all disappear if one region is removed
Task | 7685 | Bug Check validation of mandatory fields at Save as well as Activate
Task | 7686 | Flora and Fauna Profiles only - Lock Species once Activated


xzzy (approved):
Sensitivity Check Completed

PR #553, opened 2025-02-17
Better way to load CSRF_TRUSTED_ORIGINS as python list with one liner

oakdbca:


preranaandure (approved):
Sensitivity check completed.

PR #554, opened 2025-02-18
Security update

oakdbca:
- Resolve false positive dependabot alert by updating esbuild version and running npm update.


preranaandure (approved):
Sensitivity check completed.

PR #555, opened 2025-02-19
Security Update

oakdbca:
- Attempt to override version of esbuild that vite uses to resolve false positive dependabot alert


xzzy (approved):
Sensitivity Check Completed

PR #565, opened 2025-04-28
General UAT Changes

oakdbca:
- Add export system group users to .csv functionality
- Add genergic export to .csv funcionality to all django lists
- Fix django cron bug with incorrectly named parameter 'run_on_days' changed to 'run_weekly_on_days'
- Make sure that for all exporting from datatables that the untruncated value is used in the export file
- Dev ops task 8359: Communities Dashboard - Status filter needs 'Discarded' added as an option.


xzzy (approved):
Sensitivity Check Completed

PR #567, opened 2025-05-01
Uat Changes General

oakdbca:
- CS: Show related items link for species and community records only when viewing external referral CS
- Remove "No changes" warning message on SC details pages as no longer relevant
- Add better validation error when uploaded shapefile .shp has an invalid crs / no crs
- Add code to remove non .zip shapefile documents if the .zip validation fails so that only the .zip remains on the front end


xzzy (approved):
Sensitivity Check Completed

PR #569, opened 2025-05-05
UAT Change General

oakdbca:
- Add entries for all django admin csv export settings to settings.py
- Work around bug with django admin csv export action by setting default escape character to a \
- Dev ops task 8365: UX OCC Add Threat from OCR - change button text
- Dev ops task 8363: CS Dashboard Need to be able to filter by Unlocked
- Dev ops task 8396: OCR Flora & Fauna Count Method and Values
- Dev ops task 8397: OCR Flora & Fauna Count default values should be Null not 0
- Dev ops task 8398: OCC Flora & Fauna Count default values should be Null not 0
- Dev ops task 8399: OCC Flora & Fauna Count Method and Values
- npm update


xzzy (approved):
Sensitivity Check Completed

PR #576, opened 2025-05-12
UAT Changes General

oakdbca:
- Dev ops task 8401: CS - Send External Referral Invite Bug
- Dev ops task 8372: External User - Home Page - Add Outstanding Referrals List
- Dev ops task 7953: OCC Map - error message when making OCR geometries visible on map


xzzy (approved):
Sensitivity Check Completed

PR #592, opened 2025-06-30
UAT Changes General

oakdbca:
Task | 10695 | OCR Fauna Animal Observation - New Django List Field "Animal Behaviour"
Task | 10723 | OCC Fauna Animal Observation - New Django List Field "Animal Behaviour"
Task | 10724 | OCC Need a "Deactivate" Occurrence action button that changes status to "Historical"

- Bug fix: Make sure if the user is proposing to approve and assign to existing OCC that the new occurrence name is left blank.
- Change certain decimal fields to serialize as float for the presentation layer. This fixes bugs with the new dirty fields detection which was giving false positives due to conversion between string and float on the front end.
- Remove close warning on all history modals

### DIFF
--- a/boranga/components/occurrence/api.py
+++ b/boranga/components/occurrence/api.py
@@ -3874,9 +3874,9 @@ class OccurrenceViewSet(
         ],
         detail=True,
     )
-    def close_occurrence(self, request, *args, **kwargs):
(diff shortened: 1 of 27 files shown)


xzzy (approved):
Sensitivity Check Completed

PR #598, opened 2025-07-09
UAT Changes General

oakdbca:
Task | 8358 | Communities Only - change Area of Occupancy scale to 10km x 10km
Task | 10857 | UX Occurrences Dashboard: if user expanded a table, remember it is expanded when returning to dashboard
Task | 10861 | OCC Sites: WA Boundary needs to include remote islands in WA jurisdiction
Task | 10823 | CS Status "Unlocked" records must remain 'active'
Task | 10824 | CS "Unlocked" must automatically re-lock


xzzy (approved):
Sensitivity Check Completed

PR #602, opened 2025-07-14
UAT Changes General

oakdbca:
Bug fix for occurrence name lookup on all OCC dashes (was only searching active occurrences now returns all)
Bug fix for community ID lookup initialisation on OCC Communities dash
Bug fix for new lock/unlock OCC mecahnism to use PATCH when calling api end point instead of POST


xzzy (approved):
Sensitivity Check Completed

PR #604, opened 2025-07-14
UAT Changes General

oakdbca:
Task | 10849 | SC Status & Dashboard filter: display a new status combination: Historical-Private & Historical-Public
Task | 10871 | New Community Profiles - add a warning if the entering a Community Name or ID that is already in use
Task | 10879 | CS Effective From/To Dates are mandatory and needs to be formatted as such
Task | 10880 | CS Bug edits to Effective From Date are not saving


xzzy (approved):
Sensitivity Check Completed

PR #607, opened 2025-07-15
Updated devcontainer to calling the 2404 postgres load script and lau…

ka-pr:
…nch and task json files to starting vite instead of node serve


xzzy (approved):
Sensitivity Check Completed

PR #608, opened 2025-07-16
Added import from file-button to a select list of model admin classes

ka-pr:
- Adds import_export::ImportMixin to list style model admin classes
- Allows to import data into the model from a file (csv, xlsx, json, ...)
- Columns may be omitted in the file which results in blank values
- Existing data may be overwritten by providing the ids of the records in the file


xzzy (approved):
Sensitivity Check Completed

PR #609, opened 2025-07-18
Dev task 10726 10727 change referees add documents

ka-pr:
- Changed permission for referees to allow them to add and edit documents for Occurrence Reports and Conservation Statuses
- Fixed an error being thrown from a CS validate function about a missing value during draft status that is not available at that time
- Moved the uploaded_by field from the OCR/OC document model classes into their parent class: Document
- Uploaded by information is now available to all documents that inherit from Document, but is only implemented for OC/OCR (old) and Conservation Status (new)


xzzy (approved):
Sensitivity Check Completed

PR #612, opened 2025-07-25
UAT Changes General

oakdbca:
- Reworking of split species profile tab: Allowing user to copy the original values for the regions, districts, distribution and general sections
- Attempt to fix recent OCR bug that is happening in UAT for draft OCRs but difficult to replicate locally


xzzy (approved):
Sensitivity Check Completed

PR #613, opened 2025-07-28
UAT Changes General

oakdbca:
Bug fix for datatables that are client side and were showing no records
Bug fix for CS scientific name/community name not being editable as draft
Ongoing work on split species


xzzy (approved):
Sensitivity Check Completed

PR #614, opened 2025-07-28
UAT Changes General

oakdbca:
Dev ops task 11002: SC convert 'Extent of Occurrences' from an integer field to a number with 5x decimal places
Work on Species Split Front end, api and emails.


xzzy (approved):
Sensitivity Check Completed

PR #615, opened 2025-07-29
Dev task 10884 ocr to orf

ka-pr:
Renames occurrence report's OCR prefix to ORF according to dev task 10884 and provides a migration file to rename the occurrence report number in existing reports accordingly.


xzzy (approved):
Sensitivity Check Completed

PR #617, opened 2025-08-01
Dev task 10977 display common name under schientific name

ka-pr:
Adds common name lookup to the conservation list detail page and formats how they are displayed.


xzzy (approved):
Sensitivity Check Completed

PR #618, opened 2025-08-01
Changed tilelayer serializer to unescape its geoserver url that is pa…

ka-pr:
…ssed to the frontend map


xzzy (approved):
Sensitivity Check Completed

PR #620, opened 2025-08-12
Dev task 10970 10975 10991 map related tasks

ka-pr:
- Changes map svg icons to be imported as vue components, so icons are displayed regardless of their relative path
- Changes the default map zoom to 13.3.. which roughly translates to a map scale of 1:50k. This only applies to zooming to single features.
- Adds the file-upload-disabled property to ocr maps, so file upload is disabled and greyed out based on what value the parent component passes in


xzzy (approved):
Sensitivity Check Completed

PR #624, opened 2025-08-15
Community Rename

oakdbca:
- Community Rename first version of rework
- Add 'renamed to' list of communities to community details page to speed up testing
- Tweak django vite configuration
- Tweak global fetch override so it only redirects 401s to login when request path starts with /api
- Remove some unused npm packages
- npm update


xzzy (approved):
Sensitivity Check Completed

PR #626, opened 2025-08-18
Rename and Split Species Tweaks

oakdbca:
- Update rename species email template to match rework
- Adjust rename species so that it overwrites the data in the resulting species with data from the original species
- Add a scrollable div to the split species assign occurrences tab to support any number of occurrences


xzzy (approved):
Sensitivity Check Completed

PR #629, opened 2025-08-20
User Group Edge Case Bug Fix

oakdbca:
Bug fix: That was allowing people who are only in the internal contributors group to be then added to the external contributors group. Modified code to explicitly check if a user is in the internal contributors group and if so to not add them to external contributors.


xzzy (approved):
Sensitivity Check Completed

PR #631, opened 2025-08-20
Data Migration Prep

oakdbca:
Dev ops task 10997: Taxonomy all Species - rename Phylo/Phylogenetic Group as Informal Group (per Nomos term)
Migration Dev ops Task | 11151 | OCR/OCC Estimated Population Area - change from Integer to Number field
Migration Dev ops Task | 11147 | CS Flora - Rename Phylo Group in dashboard filter
Migration Dev ops Task | 11145 | SC Flora and Fauna - Rename Phylogentic Group


xzzy (approved):
Sensitivity Check Completed

PR #632, opened 2025-08-21
Dev task 10976 10978 conservation status dont hide certain form fields

ka-pr:
- Displays select conservation status fields in species&communities details view when there is no current CS (10978)
- Unhides effective to, first listed and due date fields when they are not set (10976)


xzzy (approved):
Sensitivity Check Completed

PR #634, opened 2025-08-26
Changed Highlight on Map for Occurrence Tenures and Related OCR to zo…

ka-pr:
…om to the respective feature with a max zoom of 1:50000. Also fixed both components to properly pan to the map and not to the top of the document


xzzy (approved):
Sensitivity Check Completed

PR #641, opened 2025-09-17
NOMOS Related Fixes and UAT Feedback Tasks

oakdbca:
- Change unique constraint from scientific_name to taxon_name_id in Taxonomy model
- Remove duplicate checking from nomos blob data script
- Add simple script to check nomos blob data files for duplicates in case we need in future
- When a taxonomy search result lookup returns two or more identical scientific names (very rare) add NOMOS taxon name id to the result so the user can confirm which is the correct record

UAT Feedback Tasks

- Fix regression that disabled the filtering of categories based on the selected list on CS profile pages
- Add archiving of districts to both species/communities profile and OCR/OCC profile pages


xzzy (approved):
Sensitivity Check Completed

PR #644, opened 2025-10-02
Build fix

oakdbca:
Allow .git for now as we are still copying git repo to image and usin…g to get git hash (should consider passing git commit in via the pipeline build command instead to avoid that).


xzzy (approved):
Sensitivity Check Completed

PR #645, opened 2025-10-20
Secuirty Changes & OCR Bulk Importer Fixes & Misc

oakdbca:
- Strip html tags from strings when saving models
- Randomise file name on disk for user uploaded files
- Fix issue with OCR Bulk Importer where embedded validation cause file corruption under MS Windows OS
- Make embedded list validation for bulk import schemas more robust by using hidden sheet ranges
- Fix regression with uploading files to species profile
- Fix some front end issues with 'create all' and 'create all mandatory' schema columns for a model
- Try to fix some styling issues with ocr bulk importer that are happening in uat but not locally
- Optimise spatial functions for calculating areas on species profile page
- pass git commit hash into docker image via build args (avoids copying git repo into docker image)
- npm update


xzzy (approved):
Sensitivity Check Completed

PR #647, opened 2025-10-21
Security Enhancements & Misc

oakdbca:
- Add ENV for max file upload size
- Remove and replace department_users end point that was returning all users in every request
- Fix issue where a missing file was resulting in 500 error on public species/communities profile pages
- Fix issue where loading animations did not stop if the internal profile image file was missing
- Add feature to show default value (if exists) for columns/fields in OCR Bulk Importer
- Fix bug with hashing related item when one or more of the variables to be hashed has None value
- Fix Bug in OCR Bulk importer validate function that tries to use 'get' to select sample value based on non unique field (if multiple records returned it now just chooses the first one)
- Fixed bug in bulk importer where a method was returning None instead of an empty queryset
- No longer allow fauna species profiles to select auto for Actual Area of Occupancy (value must be manually entered rather than calculated from GIS functions as fauna is based on points which have 0 area)
- Fixed GIT_COMMIT_HASH issue
- Added copilot-instructions.md file to help copilot be more helpful in this codebase
- npm update
- pre-commit autoupdate


xzzy (approved):
Sensitivity Check Completed

PR #650, opened 2025-10-31
Misc UAT Tasks and bux fixes

oakdbca:
- Dev ops task 10871: Fix issue with validation error key showing backend field name 'community_migreated_id' which doesn not match expected front end name 'Community ID'.
- Dev ops task 10860: Customise invalid point location message as requested.
- Dev ops task 12889: CS Defer pop-up - 'Reason/Comments' is mandatory and should be formatted as such.
- Dev ops task 10853: OCC automation: on change of Species or Community Name, auto change related OCRs to match.

Bug fix of CadastreLayer: Tenure area id was missing layer name prefix and owner name and owner count fields were missing from model so were not being populated to tenure area.


xzzy (approved):
Sensitivity Check Completed

PR #659, opened 2025-11-20
Misc UAT and Data Migration Prep

oakdbca:
- Dev ops task 13451: OCC View ORF Section - multi-select fields are excluded from the display
- Removed 2 redundant checks from new nomos fetch handling
- Bug fix that wasn't showing Soil Types when copying from ORF to OCC
- Add datetime_updated value when doing bulk update of archived/non currence records in nomos data handling script


xzzy (approved):
Sensitivity check completed

PR #670, opened 2026-01-13
UAT Feedback & Data Migration

oakdbca:
Task | 13903 | CS Fix Permissions re Unlock to Edit & use the Back to Assessor action
Task | 13874 | Bulk ORF Schema Validation Error needs to be permitted for schema but not per row
Task | 13901 | CS Approval ActionLog - missing comment plus has an extra misleading log stating it was Closed
Task | 13900 | CS: when an approved or inactive CS is unlocked, date fields should be editable
Task | 13878 | CS - Make Name (Scientific Name/Community Name) editable only when status is Draft or With Assessor
Task | 13897 | CS Completed Referee action icon (email) is confusing - change to a repeat icon
Task | 13898 | CS Assigned Assessor not displaying

Plus signifigant work on TPFL and TEC legacy data migration code.


xzzy (approved):
Sensitivity Check Completed

PR #687, opened 2026-02-06
Data Migration

oakdbca:
Dev ops task 13904: CS External Contributor unable to view submitted CS Proposal (Caused by queries returning duplicate records instead of unique records. Fixed several similar issues.)
Dev ops task 13992: Make sure when saving the OCR after it's species has changed to create a new version with django reversion


xzzy (approved):
Sensitivity Check Completed

PR #690, opened 2026-02-10
Data Migration & UAT Fixes

oakdbca:
- Rework some of the CS adapter and handler to match other migration structures
- Implement any missing transformations to prepare for dev data verification

Bug fixes for:

13979 | CS Delisting change 3/5: introduce a condition (based on Change Type) to alter the final actions displayed 
13980 | CS Delisting change 4/5: validation required when Change Type = DeList is selected


xzzy (approved):
Sensitivity check Completed

PR #706, opened 2026-03-12
Bug Fixes

oakdbca:
Move from GET to POST now that OCR Datatable has lots of filtering as it was broken with the url being too long
Various fixes to make the map loading more error tolerant
Clean up unused field out of GeoserverURL model
Update fixture to remove unused field and add correct WMTS version


xzzy (approved):
Sensitivity Check completed

PR #709, opened 2026-03-19
UAT Feedback Tasks

oakdbca:
Dev ops task 15916: Multi-select filters
Dev ops task 15542: Geometry Info Cards: bug fix so that info toast shows for Sites
Dev ops task 15621: Make sure all circular polygons are generated using the same consistent method
Dev ops task 16053: ORF Approval creates new OCC: exclude documents from the ORF data copied to new OCC
Dev ops task 16052: ORF: exclude documents from the Copy ORF action

Change child rows to render fields horizontally instead of vertically to save vertical screen space when users are viewing multiple child rows


xzzy (approved):
Sensitivity Check Completed

PR #713, opened 2026-03-27
Resolve Cadastre Layer Resource Limitations

oakdbca:
- Imrove application of GDAL_CACHEMAX
- Keep the downloaded layer in the private-media PVC instead of within /app


xzzy (approved):
Sensitivity Check Completed

PR #714, opened 2026-03-30
Data Migration & UAT Feedback

oakdbca:
- Added .md file with commands to do a full run with one command
- Fix email field validation in django admin when saving ExternalContributorBlacklist
- Make sure to populate submitter information name from emailuser when importing CS
- Convert external ORF datatable to use POST instead of GET due to the querystring being too long


xzzy (approved):
Sensitivity Check Completed

PR #722, opened 2026-04-07
Bug fix

oakdbca:
Modify the fallback logic to only retry saving the rows from the specific batch that failed.


xzzy (approved):
Sensitivity Check Completed

PR #724, opened 2026-04-08
Data Migration & Minor Bug Fixes

oakdbca:
Dev ops task 14906: Model: OccurrenceGeometry - Field Name: geometry. Bug fix: Accidentally added a 1m buffer to every geometry. not required as they are already 1m circle polygons.
Dev ops task 11457: Model: AssociatedSpeciesTaxonomy - Field Name: occurrence_report. Bug fix: no mappings were working as it was using the key with the prefix e.g. 'tpfl-<migrated_from_id>' instead of just the migrated_from_id.

Other minor fixes.


xzzy (approved):
Sensitivity Check Completed

PR #725, opened 2026-04-09
Data Migration Fixes

oakdbca:
Dev ops task 16470: Bug: Tenure subrecords are only being generated for Active OCCs, but should also be generated for Historical OCCs
Dev ops task 14938: Model: OCCLocation - Field Name: region. Bug fix: Remove unnecessary mapping and populate district id in adapter extract so that pipeline has correct value to work with
Dev ops task 14935: Model: OCCLocation - Field Name: locality. Bug fix of column mapping name
Dev ops tasks 14912, 14916, 14936, 14939, 14940, 14941: Bug fix: Correct the column mapping names so values actually make it to pipelines
Dev ops task 14906: Model: OccurrenceGeometry - Field Name: geometry. Bug fix: Accidentally added a 1m buffer to every geometry. not required as they are already 1m circle polygons.


xzzy (approved):
Sensitivity Check Completed

PR #737, opened 2026-04-24
Data Migration Fixes

oakdbca:
- Fix transform of cadastre features so that they go through the gda94 to gda2020 transform (which is not done correctly when going from WGA84 to gda2020 like it was before)
- Fixes to associated species taxonomy to avoid orphan records and make sure fresh records are created each time rather than reusing incorrect records
- Fix for datetime_updated default value (use datetime_created if no legacy value)
- Fix for last_modified_by (use submitter if no legacy value)


xzzy (approved):
Sensitivity Check Completed

PR #738, opened 2026-04-28
Final TPFL Data Migration Changes

oakdbca:
- Add OCCLocation and OCRLocation models to history seeder
- Add missing last_modified_by field to OccurrenceRow to fix missing values
- Replace feature id (fid) that comes from geoserver with CAD_PIN as a cadastre tenure identifier
- Fix bug that was preventing the info tool from showing for valid data layers


xzzy (approved):
Sensitivity Check Completed

PR #739, opened 2026-04-28
Temporarily remove reference to CAD_OWNER_NAME as field is missing from PROD cadastre layer

oakdbca:


xzzy (approved):
Sensitivity Check Completed

PR #759, opened 2026-05-28
UAT Tasks

oakdbca:
Dev ops task 18273: Historical Profiles should still be accessible to the Public/External Contributors IF the Published status = Historical.
Dev ops task 18271: Replace lock Geometry with Ability to Discard and Reinstate Geometries
Dev ops task 18268: ORF Dashboard Table fields - Add "Suggested Occurrence Name"
Dev ops task 16496: Bulk Import Flora and Fauna - auto-calculate value for ORFNUM Count Status. Add an update in the post processing stage (Requi
red as the save method is programmed to not change count values for records that have a migrated_from_id for protecting values during the legacy 
import process)
Dev ops task 18266: Bulk ORF Importer: Bug With Assessor ORF assigned to New OCC [OccurrenceReportApprovalDetails] is creating the new OCC. Resol
ved by removing Occurrence from model to create unless the processing status is approved


xzzy (approved):
Sensitivity Check Completed

PR #762, opened 2026-06-02
Pin select2 version to avoid incompatibility with jquery 4.

oakdbca:
Pin select2 version to avoid incompatibility


xzzy (approved):
Sensitivity Check Completed

PR #778, opened 2026-07-03
Data Migration and UAT Tasks

oakdbca:
Dev ops task 18402: Add new public homepage section help text
Dev ops task 18642: Make sure the  front end passes the occurrence id in the occurrence_id property rather than just occurrence so the serializer is happy
Dev ops task 12291: Fix code that copies ORFAssociatedSpecies comment to cloned OCC

Plus other misc tasks


preranaandure (approved):
Sensitivity check completed.

PR #779, opened 2026-07-03
UAT Tasks

oakdbca:
Tasks so that users in query user group can't see actions they shouldn't and so that forms are readonly/disabled as appropriate


xzzy (approved):
Sensitivity Check Completed

PR #780, opened 2026-07-06
UAT Tasks

oakdbca:
- Make sure custom Boranga PRIS test on accounts page is retrieved from the DB so it updates when the users update it in django admin
- Tweak default run time of cadastre layer copy so that it doesn't overlap with the NOMOS sync
- make the location input disabled when in read only states since the readonly attribute alone still looks editable
- Allow specifying the title property as for the minutes view document modal the property is minutes_number rather than the usual document_number


preranaandure (approved):
Sensitivity check completed.

PR #781, opened 2026-07-06
UAT Tasks

oakdbca:
- Tweak horizontal space available for the occurrence reports epsg code and datum so ideally the full text is visible
- Roll out the new status and role matrix pattern to the meeting details page.
- Use dynamic title property for document display in view_document component
- Further improvements to the new status and role based permission matrix pattern example used in the meetings component


xzzy (approved):
Sensitivity Check Completed

PR #783, opened 2026-07-08
Data Migration and Bulk Import Tasks

oakdbca:
- Massive performance enhancements of history cleaner and reseeder
- bulk ORF importer:
  - Bug fix: Improve processing time estimates
  - Bug fix: Fix regression that was preventing items from being queued quickly
- Added management command to clean out orphaned django reversion revisions (i.e. those with no associated Version obejcts)
- Add cache of retrieve bulk import schema as large schemas like 100 columns and over can take 30+ seconds to retrieve from the api. Once cached they will take 150ms unless the schema (or one of it's columns) has been changed.


xzzy (approved):
Sensitivity Check Completed

PR #784, opened 2026-07-09
Data Migration & Bulk Import Tasks

oakdbca:
- Fix to bulk importer to stop it jumping straight to processing queue item before the pre processing step can complete
- Add logging to reversion seeder so there is feedback on progress to show the process hasn't stalled


xzzy (approved):
Sensitivity Check Completed

PR #785, opened 2026-07-13
Data Migration

oakdbca:
- Make sure to store original ewkb with original datum.
- Update the tfauna point transformation to interpret all incoming legacy data as GDA94 (Igore the DATUM column)
- Update front end packages


xzzy (approved):
Sensitivity Check Completed

PR #786, opened 2026-07-14
Update ows query layer name to new cadastre ownership layer

oakdbca:
Update ows query layer name to new cadastre ownership layer


xzzy (approved):
Sensitivity Check Completed

PR #787, opened 2026-07-14
Data Migration

oakdbca:
- Add proper support for --filter-ids flag to tfauna occurrence tenure run so I can retry ones that are failing
- Add logging to tally up how many intersects are missing as the migration runs for better diagnostics
- Add code to be able to raise an error if local cadastre is available but has a transient failure


xzzy (approved):
Sensitivity Check Completed

PR #802, opened 2026-07-24
Miscellaneous

oakdbca:
- Add any last data migration code that was not previously commited (can be removed at a later date)
- Include required development packages in dev container image
- Upgrade github action versions to avoid deprecation, improve performance and security
- Upgrade front end packages


preranaandure (approved):
Sensitivity check completed.

PR #803, opened 2026-08-03
Tasks and Bug Fixes

oakdbca:
- Dev ops task 18762: Fix an issue that can easily cause the creation of duplicate geometries (which can compound) Once there are duplicate geometries on
-  the map this can cause confusing behaviour for users
- Dev ops task 18563: First version of new slimmed down KB reports
- Dev ops task 18808: Apply some fixes to the Edit Site modal. So that the focus works and degrades gracefully and the Occurrence Reports data is loaded 
- correctly.
- Dev ops task 18804: Add the latitude and longitude for the point geometry for these 8 queries as requested
- Dev ops task 18804: Add the observation method field to the 13 queries requested
- Dev ops task 18804: REMOVE animal_observation_detail_comment (KB Name AN_OBS_COM) from Boranga_FaunaOCCPointsActive

Plus misc fixes and reporting

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,7 @@ COPY --chown=oim:oim scripts/combine_csvs.py \
                       scripts/dedupe_errors.py \
                       scripts/generate_uat_fixtures.sh \
                       scripts/partition_migration_data.py \
+                      scripts/repport_duplicate_geometries.py \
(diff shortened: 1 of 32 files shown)


preranaandure (approved):
Sensitivity check completed.

PR #805, opened 2026-08-06
Fix: Rename, split and combine db performance

oakdbca:
Rename, split and combine all suffer timeouts if the underlying species has 10s of thousands of related OCCs / OCRs.

This change fixes the issue by:

- Converting those end points to use bulk insert / update for all DB operations
- Fix some issues with the loading state of front end buttons so clearly show the buttons are disabled and have a loading spinner
- Warn the user in the modal that rename, split or combine can take up to one minute for records with lots of relations

Misc

- Remove dead unused cache key delete lines for CACHE_KEY_MAP_OCCURRENCES


preranaandure (approved):
Sensitivity check completed.

PR #806, opened 2026-08-06
Dev ops task 18833: Allow greater than and less than to render in text inputs

oakdbca:
- Use nh3.clean_text instead of nh3.clean which was overkill
- Make sure all datatables escape html and render the plain string instead


preranaandure (approved):
Sensitivity check completed.

PR #807, opened 2026-08-06
Dev ops task 18804: KB Spatial Reports - Field Changes

oakdbca:
Further changes that the business users have requested in the meantime.


preranaandure (approved):
Sensitivity check completed.

PR #808, opened 2026-08-07
Fix regression caused by datatable html sanitisation

oakdbca:
fix: The previous change to render was clobbering the fixed actions column styling. This way achieves the same thing without doing that.


preranaandure (approved):

preranaandure (approved):
Sensitivity check completed.

PR #811, opened 2026-08-11
Dev ops task 18842: Fix the display of utf special characters in csv and xlsx reports

oakdbca:
- Encode csv file with utf-8-sig instead of just utf-8
- For xlsx bypass the str function and all files directly to allow openpyxl to deal with the types appropriately


preranaandure (approved):
Sensitivity check completed.

PR #816, opened 2026-08-18
fix: SQL syntax error

oakdbca:


xzzy (approved):
Sensitivity Check Completed

PR #817, opened 2026-08-19
Fix/18879 tenure deduplication

oakdbca:
This change fixes the tenure duplication bug identified by the business users in a way that preserves all user inputed data of records that are already affected whilst preventing the same from happening in future.


xzzy (approved):
Sensitivity Check Completed

PR #818, opened 2026-08-20
Fix: Related Items Actions Link Not Rendering as HTML

oakdbca:
Do not escape html for column with data attribute 'action_url' as that contains safe html action links for the related items pages


xzzy (approved):
Sensitivity Check Completed

PR #819, opened 2026-08-20
Fix the sorting on number of reports

oakdbca:


xzzy (approved):
Sensitivity Check Completed

